PR #22, opened 2024-09-22
Initialize crate structure

alice-i-cecile:
To ease collaboration and development, I've created stub crates for all of the crates in the current milestone. These are quite granular, to make sure that users can use only the bits that they want for various functionality in their own tools and games.

I'm sure we'll need more eventually, but this is a good place to start!

I'm following `bevyengine/bevy`'s project structure to ease the pain if and when we merge this into the same repo.


ChristopherBiscardi (approved):
lgtm, tests currently pass and build works.

Do we have a suggested "widget" abstraction at the moment or are we waiting for constructs/bsn for that?

Each individual widget should probably have its own usage example too, similar to the examples in the bevy repo. Maybe just a note in the README for later or a single scaffolded example with a 2d camera.

alice-i-cecile:
> Do we have a suggested "widget" abstraction at the moment or are we waiting for constructs/bsn for that?

I'm going to hack something in for now. I'll probably base it off the sickle_ui patterns for now.

> Each individual widget should probably have its own usage example too, similar to the examples in the bevy repo.

Fully agree!

PR #23, opened 2024-09-22
Change bevy dependency to be main branch of bevy & Reserve components for Pane Layout system

Atlas16A:


alice-i-cecile (changes requested):
I think I'm okay to swap to bevy main for now (LWIM is the only immediately planned third party dependency) but we need to pick a specific arbitrary commit to avoid uncontrolled breakage.

bushrat011899:
> I think I'm okay to swap to bevy main for now (LWIM is the only immediately planned third party dependency) but we need to pick a specific arbitrary commit to avoid uncontrolled breakage.

I'm not familiar with how easy it would be to setup, but perhaps a something like [dependabot](https://docs.github.com/en/code-security/dependabot) could be used to automate _trying_ to keep the editor aligned with `bevy/main`?

Atlas16A:
I mustve done something wrong cause github is automatticly making my commits combine into this PR

Atlas16A:
Ill make a new PR and seperate these commits

PR #25, opened 2024-09-23
Added Asset store crate

Atlas16A:
Adds asset store crate with async-stripe crate built in to be ready to begin implementing the asset store.

### DIFF
--- a/crates/bevy_asset_store/Cargo.toml
+++ b/crates/bevy_asset_store/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "bevy_asset_store"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-stripe = "0.39.1"
--- a/crates/bevy_asset_store/Cargo.toml
+++ b/crates/bevy_asset_store/Cargo.toml
@@ -4,4 +4,3 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-async-stripe = "0.39.1"
--- a/crates/bevy_asset_store/src/lib.rs
+++ b/crates/bevy_asset_store/src/lib.rs
@@ -1,0 +1,16 @@
+//! A asset store for Bevy Editor.
+
+pub fn add(left: u64, right: u64) -> u64 {
+    left + right
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn it_works() {
+        let result = add(2, 2);
+        assert_eq!(result, 4);
+    }
+}


crates/bevy_asset_store/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "bevy_asset_store"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-stripe = "0.39.1"
```
bushrat011899, line 7: I feel like this might be a controversial inclusion. Artists should definitely be paid for their work, but I don't know if the Bevy Foundation wants to be involved in developer-artist payment processing. I think it would be a better focus for the editor to work with user configurable asset stores (e.g., a local folder, a remote server using some REST API, etc.). If the Bevy Foundation wants to setup a paid store at some point in the future then the editor wont need any additional code changes to support it.

crates/bevy_asset_store/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "bevy_asset_store"
```
bushrat011899, line 2: Should the asset store be limited to just assets? Unity and Unreal can also provide code items as a part of their stores, which for Bevy would likely mean plugins.

bushrat011899 (changes requested):
I do think the editor should include a way to browse local and remote assets for inclusion into your project, but I feel wary about building in payment processing into the editor. The multinational nature of Bevy, terms of service with Stripe, etc. are all cans of worms that might not be worth it.

Atlas16A:
> I do think the editor should include a way to browse local and remote assets for inclusion into your project, but I feel wary about building in payment processing into the editor. The multinational nature of Bevy, terms of service with Stripe, etc. are all cans of worms that might not be worth it.

To be clear, this will just be for handling the connection between the editor and bevy foundation and additional visuals that the asset browser won't have (like a price tag widget, etc), I added stripe because I wasn't sure weather or not it needed to be on the client side to handle payments correctly.

According to Alice bevy foundation does plan to have an asset store as a way for users and bevy to have income outside of donations, similar to other editor stores.

bushrat011899:
> According to Alice bevy foundation does plan to have an asset store as a way for users and bevy to have income outside of donations, similar to other editor stores.

Ok that's me out of the loop on those plans, thanks for clarifying!

alice-i-cecile:
We should avoid handling payments in the editor itself: we should open a web page instead for that. That way, we don't need to worry nearly as much about security for the editor or extensions.

Atlas16A:
> We should avoid handling payments in the editor itself: we should open a web page instead for that. That way, we don't need to worry nearly as much about security for the editor or extensions.

That makes sense, I wasn't sure how we'd manage it.

alice-i-cecile:
For previous discussion on this context, see https://github.com/bevyengine/bevy-website/issues/175.

The web-based storefront will grow out of https://bevyengine.org/assets/ :)

crates/bevy_asset_store/Cargo.toml
alice-i-cecile: Yeah, IMO swapping this to `bevy_marketplace_viewer` or something is more clear.

crates/bevy_asset_store/Cargo.toml
alice-i-cecile: Yep, as discussed this should be cut: we'll let the webstore handle payments.

crates/bevy_asset_store/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+//! A asset store for Bevy Editor.
```
alice-i-cecile, line 1: I think we should have a clearer vision of how the editor interacts with the store, and what this crate handles.

I'd be interested in functionality for checking for updates to assets, and breadcrumbs to the website to browse assets, but at least for now I'd like to keep the data and interface totally on the website to avoid having to duplicate efforts.

Atlas16A: My personal interpretation is the crate acts as a link between the website asset store and the editor's asset browser pane, implementing things like loading previews of assets, essential it'd be full of networking code to connect the editor to the store. I figured it best to be separated from the asset browser so if someone wants they can easily cut out the asset store no problem.

alice-i-cecile: Hmm, I see! Can you add a `bevy_asset_preview` crate as part of this PR and explain the relationship between the two in the module docs here?

rewin123:
Since we now have milestones corresponding to the roadmap, I took the liberty of assigning a milestone to this PR. Feel free to change the assigned milestone, as I don't have authority here.

Atlas16A:
Closing Pr to redo my fork

Issue #29, opened 2024-09-23
Periodically update to the latest Bevy `main` commit


alice-i-cecile:
I think that tracking `main` is currently more useful than latest stable, in order to get and test new UI features ASAP.

However, we need to pin a specific commit, rather than just the `main` branch, to avoid getting sudden failures due to upstream breaking changes (and bugs).

Manually bumping this is a chore, but we can automate this process with a daily job. That job should open a PR (or bump its most recent unmerged PR), which can be tested, reviewed and merged.

rewin123:
Since we now have milestones corresponding to the roadmap, I took the liberty of assigning a milestone to this PR. Feel free to change the assigned milestone, as "I have no power here"

alice-i-cecile:
Fixed in #43.

PR #43, opened 2024-09-26
Implement gh action for updating bevy periodically

no-materials:
Addresses #29 

The cron job is run weekly every Monday midnight UTC, or can be triggered manually.

Action flow:

* Fetches the latest bevy main rev
* Logs in as the `github-actions` bot user, makes the replacement in `Cargo.toml`, commits changes
* Checks for an existing PR with the branch name `update-bevy` and the author name containing `github-actions` (exact matching here didn't quite work, not sure why)
* If the existing PR is found, it bumps it, or else a new PR is created

### DIFF
--- a/.github/workflows/update-bevy.yaml
+++ b/.github/workflows/update-bevy.yaml
@@ -2,7 +2,7 @@ name: Update Bevy Dependency
 
 on:
   schedule:
-    - cron: "0 0 * * *" # Runs daily at midnight UTC
+    - cron: "0 0 * * 1" # Every Monday at 00:00 UTC
   workflow_dispatch: # Allows manual trigger as well
 
 jobs:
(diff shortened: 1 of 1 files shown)


Atlas16A:
I was thinking weekly? Daily seems like a lot for making sure devs keep pulling latest, in my opinion.

no-materials:
> I was thinking weekly? Daily seems like a lot for making sure devs keep pulling latest, in my opinion.

You def have a point, daily is to extreme I guess :sweat_smile: Changed it to weekly (randomly selected Monday midnight)

Atlas16A:
Monday seems fine to me, that way if something breaks its not on a weekend or smth.

alice-i-cecile (approved):

alice-i-cecile:
Awesome, let's try this out :)

PR #53, opened 2024-10-01
Explain why the Bevy editor can't just be a plugin for existing tools

alice-i-cecile:
Prompted by discussion on [Discord](https://discord.com/channels/691052431525675048/789987800014651412/1290697160907817072).

### DIFF
--- a/design-book/src/design-constraints.md
+++ b/design-book/src/design-constraints.md
@@ -39,7 +39,8 @@ Below each bullet point, motivation is given.
 
 - The editor is not an asset or code creation tool (certainly not in the default configuration). If an established tool does a good job authoring content, focus on importing its output, not reinventing Blender, VSCode, Asesprite, and Reaper.  
   - We have limited resources, and this isn't a good use of them. Artists and professionals already know and prefer specialized tools.
-- The Bevy editor should not have a native look-and-feel, and should not use native menus.
+  - Instead, the Bevy Editor is a scene creation tool with powerful Bevy-specific debugging capabilities, with limited asset-tweaking capabilities for quick and simple tasks.
+- The Bevy Editor should not have a native look-and-feel, and should not use native menus.
   - This is a poor use of resources, leads to inconsistent aesthetics, and makes it harder to test and teach across platforms. Instead, focus on the useful behavior and design conventions.
   - File picking widgets are an exception: these are non-intrusive, good crates exist and are not central to the user experience.
 - End users should not have to install or update Rust.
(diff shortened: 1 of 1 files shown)


AmitHossain:
It is very rare that games of any significant scope are made without designers, artists, animators, and/or other asset creators. An editor is a core part of game engines. By including an editor by default, it would empower asset creators to use the engine, rather than just rely on the programmers. It would also make Bevy less intimidating for non-programmers, allowing for the growth and proliferation of the Bevy engine amongst game devs, and thus lead to more contributors helping build the engine, both in terms of numbers and diversity of backgrounds/experiences.

alice-i-cecile:
> It is very rare that games of any significant scope are made without designers, artists, animators, and/or other asset creators. An editor is a core part of game engines. By including an editor by default, it would empower asset creators to use the engine, rather than just rely on the programmers. It would also make Bevy less intimidating for non-programmers, allowing for the growth and proliferation of the Bevy engine amongst game devs, and thus lead to more contributors helping build the engine, both in terms of numbers and diversity of backgrounds/experiences.

Ah, as an argument for why this shouldn't just be a plugin? I can get behind that. Let me try and make this more succinct.

CalculatedToast:
> > It is very rare that games of any significant scope are made without designers, artists, animators, and/or other asset creators. An editor is a core part of game engines. By including an editor by default, it would empower asset creators to use the engine, rather than just rely on the programmers. It would also make Bevy less intimidating for non-programmers, allowing for the growth and proliferation of the Bevy engine amongst game devs, and thus lead to more contributors helping build the engine, both in terms of numbers and diversity of backgrounds/experiences.
> 
> Ah, as an argument for why this shouldn't just be a plugin? I can get behind that. Let me try and make this more succinct.

I can be a bit verbose. 😅 

If not technically, then at least philosophically, a plugin implies that something is optional and not a core part of the engine (at least from my non-programmer background). I believe that an editor _should_ be core to the engine, as pretty much all non-programmer game devs cannot work without it.

Atlas16A (approved):

design-book/src/design-constraints.md
```diff
@@ -53,6 +54,14 @@ Below each bullet point, motivation is given.
   - Dogfooding and improving `bevy_ui` is a major goal. If something sucks, fix it upstream!
 - No scripting languages.
   - This targets a very niche audience: Rust is safer and easier to write than C++ and pure artists won't write Lua either. Prefer GUI controls, and then defer to node-based logic for highly complex art/design workflows.
+- The Bevy Editor is not a plugin for a third-party editor like Blender or LDTK. While tools like [Blenvy](https://github.com/kaosat-dev/Blenvy) can be very productive for a team making a specific game, they're not a replacement for the Bevy Editor.
+  - The existence of a standalone Bevy Editor tool (even with relatively limited capabilities) is valuable for helping new users get comfortable with the idea of making games in Bevy and become a functional member of a team.
```
n1ght-hunter, line 58: i would possibly say here that the editor will be extendable with plugins same as bevy. to off shoot limited capabilities

MiniaczQ:
Using "plugin" here is a bit confusing, for a second I thought bevy-editor would require an alternate method of connecting to bevy apps that isn't a plugin

Cedar-81:
Something to consider is the Next Generation Scene Ui proposition that is in the works, to take full advantage of that and show its capabilities bevy needs its own standalone editor that is fully customized for what innovations bevy decides to bring about in the future. Using a different existing tool will lead to a lot of scrambling to get things to work

design-book/src/design-constraints.md
alice-i-cecile: Yeah, that's listed in the roadmap under uncategorized work :)

PR #55, opened 2024-10-01
Add Design Guidelines

Atlas16A:
These are very very influenced by blender's guidelines, especially the formatting/structure, the goal is to get all design work on the same page and honor bevy's goals while giving users a good experience. 

Please do critique/comment/improve these docs.

PR #56, opened 2024-10-02
Add more design guidelines

alice-i-cecile:
This is a follow-up building on #55. I think that focusing on the use of the editor for teams is important, and I have a few other things to say about interaction paradigms :)

### DIFF
--- a/design-book/src/design-paradigms.md
+++ b/design-book/src/design-paradigms.md
@@ -11,6 +11,14 @@ That means:
 2. The Editor must be customizable to meet the unique needs of each user and project.
 3. Art, audio, writing and design-centric workflows should require a minimal level of technical expertise.
 
+## Redundant interaction paradigms for accessibility
+
+1. Anything that can be done via a mouse click can be done via a hotkey.
+2. Anything that can be found in a menu can be found in the command palette.
+3. Anything that can be configured via setting a menu option can be done by editing a config file.
+
+To be completely clear: the same should be true in the opposite directions!
+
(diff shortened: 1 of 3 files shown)


Atlas16A (approved):

design-book/src/design-paradigms.md
```diff
@@ -1,8 +1,33 @@
...
+1. Anything that can be done via a mouse click can be done via a hotkey.
+2. Anything that can be found in a menu can be found in the command palette.
+3. Anything that can be configured via setting a menu option can be done by editing a config file.
+
+To be completely clear: the same should be true in the opposite directions!
+
 ## Non-Overlapping
 
-The Ui should enable you to always view all relevant information and tools at any given moment, you should never have to move a window out of your way, for that reason, as well as sticking with bevy's modular structure, Bevy uses a divided window layout we call Panes.
+The UI should enable you to always view all relevant information and tools at any given moment, you should never have to move a window out of your way, for that reason, as well as sticking with bevy's modular structure, Bevy uses a divided window layout we call Panes.
```
crvarner, line 31: I don’t really understand this section below. What do you mean by a Screen being “the whole window” and multiple editors? And Panes being containers of “an” editor?

I thought the window was “the” editor and it would contain multiple panes such as an entity outliner, viewport, command palette, etc.

alice-i-cecile: Agreed, I think this terminology is unclear.

PR #58, opened 2024-10-02
Fix sub-crate dependency issue

### LINKED ISSUE #57: Sibling crates cannot be resolved on import
I noticed that when doing the following

1. Clone repo
2. Add `bevy_editor_camera` to `bevy_editor`s `Cargo.toml`
3. Import `bevy_editor_camera::*` in the `bevy_editor/src/main.rs` file
4. `cargo run` from the root path of the project

> [!NOTE]
> This happens with any sibling crate not just `bevy_editor_camera`.

the project fails to run due to unresolved dependencies.

```
error[E0432]: unresolved import `bevy_editor_camera`
  --> crates/bevy_editor/src/main.rs:15:5
   |
15 | use bevy_editor_camera::*;
   |     ^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `bevy_editor_camera`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `bevy_editor_prototypes` (bin "bevy_editor_standalone") due to 1 previous error
```

If you instead `cargo run` from within the `bevy_editor` crate itself the unresolved dependency issue is gone.
The issue here is that LSP servers still can't resolve the dependencies since their root is the workspace root.

So this seems like a configuration issue of some sort.

Dokkae6949:
Fixes #57


MiniaczQ (changes requested):
That's the wrong way of fixing it.

The problem is that workspace `Cargo.toml` tries to double as a lib/bin, which prevents members from being viable bins.

I'll post a PR for this in a second

Dokkae6949:
Does it even make a difference if the binary entry is in `crates/bevy_editor/src` or in `/src`?
I'm thinking that maybe putting it in crates makes sure that we have the future ability to completely remove the bin entry point and just use it as a lib.

MiniaczQ:
> Does it even make a difference if the binary entry is in `crates/bevy_editor/src` or in `/src`? I'm thinking that maybe putting it in crates makes sure that we have the future ability to completely remove the bin entry point and just use it as a lib.

It's more streamlines this way, if we decide to have multiple binaries it's also easier to work like this

Dokkae6949:
Might as well close this for now then.

PR #59, opened 2024-10-02
Remove lib/bin-like sections from workspace Cargo.toml

### LINKED ISSUE #57: Sibling crates cannot be resolved on import

MiniaczQ:
Fixes #57 while still keeping everything as workspace members.
Alternative to #58

Added `bevy_editor_camera` dependency in `bevy_editor` and ran `cargo run` to make sure it works.
When/if we get more binaries we'll need to use the `--bin` flag

### DIFF
--- a/crates/bevy_editor/src/main.rs
+++ b/crates/bevy_editor/src/main.rs
@@ -15,6 +15,7 @@ use bevy::prelude::*;
 use bevy_editor_camera::*;
 
 fn main() {
+    add(1, 2);
     App::new()
         .add_plugins(DefaultPlugins)
         .add_systems(Startup, camera_setup)
--- a/crates/bevy_editor/src/main.rs
+++ b/crates/bevy_editor/src/main.rs
@@ -12,10 +12,8 @@
 //! - Finally, it will be a standalone application that communicates with a running Bevy game via the Bevy Remote Protocol.
 
 use bevy::prelude::*;
-use bevy_editor_camera::*;
 
 fn main() {
-    add(1, 2);
     App::new()
         .add_plugins(DefaultPlugins)
         .add_systems(Startup, camera_setup)
--- a/crates/bevy_editor/src/main.rs
+++ b/crates/bevy_editor/src/main.rs
@@ -12,6 +12,7 @@
 //! - Finally, it will be a standalone application that communicates with a running Bevy game via the Bevy Remote Protocol.
 
 use bevy::prelude::*;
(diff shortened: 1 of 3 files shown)


Dokkae6949 (approved):

crates/bevy_editor/src/main.rs
```diff
@@ -12,8 +12,10 @@
 //! - Finally, it will be a standalone application that communicates with a running Bevy game via the Bevy Remote Protocol.
 
 use bevy::prelude::*;
+use bevy_editor_camera::*;
 
 fn main() {
+    add(1, 2);
```
Atlas16A, line 18: remove the test func you did here

Atlas16A (changes requested):

crates/bevy_editor/src/main.rs
MiniaczQ: Gotta keep it, otherwise we have unused imports (bevy_editor_camera)

MiniaczQ: Or i can do a #[allow(...)]

Dokkae6949: I suppose you can just remove the import as well since it's been tested and works now anyways. No need to keep it

Atlas16A (approved):

Dokkae6949 (approved):

PR #65, opened 2024-10-03
Dev/start on editor settings

n1ght-hunter:
here's a start on bevy_editor_settings.

i have followed what was talked about in discord and think this is not bad start.
this currently doesn't allow anyway for deep merging of project settings (required for merging the user,workspace and default settings) and also doesnt have anyway for a plugin to get or edit settings.

what it does add is 
- three forms of settings (user,workspace and default) and a way to load them from each location.
- saving for the user settings
- base models for all three types of settings. 
- updated [readme](https://github.com/bevyengine/bevy_editor_prototypes/blob/a37339c02aa7cb4cf56e95b48c7b02508b9f0911/crates/bevy_editor_settings/READEME.md)

current alternative is to directly pull in [bevy_basic_prefs](https://github.com/viridia/bevy_basic_prefs) and work from there

### DIFF
--- a/crates/bevy_editor_settings/READEME.md
+++ b/crates/bevy_editor_settings/READEME.md
@@ -1,1 +1,23 @@
 # Bevy Editor Settings
+
+This crate provides a way to have settings for the editor.
+There are three types of settings:
+- User settings
+- Workspace settings
+- Default settings
+
+## User settings
+User settings are settings that are specific to the user. They are stored in the porject at the top level in `user.toml`
+these will be moved when the editor is a standalone application
+
+## Workspace settings
+Workspace settings are settings that are specific to the workspace. They are stored in the project at the top level in `Bevy.toml`
+
+## Default settings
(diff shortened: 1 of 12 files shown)


crates/bevy_editor_settings/READEME.md
```diff
@@ -0,0 +1,23 @@
+# Bevy Editor Settings
+
+This crate provides a way to have settings for the editor.
+There are three types of settings in order of precedence:
+- User settings
+- Workspace settings
+- Default settings
+
+## User settings
+User settings are settings that are specific to the user. They are stored in the porject at the top level in `user.toml`
```
MiniaczQ, line 10:
```suggestion
User settings are settings that are specific to the user. They are stored in the project at the top level in `user.toml`
```

n1ght-hunter: ya there were a bunch of typos fixed now though

PR #77, opened 2024-10-10
add functionally to load preferences/override from a toml file

n1ght-hunter:
## Example data structure
```rs
    #[derive(Debug, Clone, PartialEq, Eq, Reflect)]
    enum AwesomeOptions {
        One,
        Two,
        Three,
    }

    #[derive(Debug, Clone, PartialEq, Eq, Reflect, Resource)]
    struct AwesomePeople {
        pub name: String,
        pub age: u32,
    }

    #[derive(Debug, Clone, PartialEq, Eq, Reflect, Resource)]
    #[reflect(@SettingsType::Project, @SettingsTags(vec!["basic", "settings", "awesome"]))]
    struct AwesomeSettings {
        #[reflect(@MergeStrategy::Append)]
        pub options_list: Vec<AwseomeOptions>,
        pub people: Vec<AwesomePeople>,
    }
 ```
### Equivalent toml
```toml
# struct with the key being the snake case variant of the type
[awesome_settings]
# array of unit enums
options_list = ["One", "Two", "Three"]
# ether this for vec of structs
people = [
    { name = "bob", age = 31},
]
# or this
[[awesome_settings.people]]
name = "joe"
age = 42
```

what is currently de serializable 
```toml

# basic struct
[basic_settings]
name = "bevy_editor_settings"


# list with replace
[list_testing]
list = ["three", "four"]

# list with append
[list_testing_append]
list = [3, 4]

# top level enum
[enum_testing]
variant = "Two"

# enum in a struct
# and enum tuple
[enum_settings]
test1 = {Tuple = ["hello", 42]}

#  a struct enum on a struct
[enum_settings.test2.Struct]
name = "four"
age = 4

# a struct with a list of enums
[enum_settings_list]
settings = ["Three"]

# top level tuple struct
# adds the fields field to the tuple sturct. this is ONLY done for a top level tuple struct
[tuple_struct]
fields = [2, "two"]

# tuple struct in a struct
[struct_with_tuple]
tuple = [3, "three"]
```

## Settings types
```rs
/// Annotation for a type to show which type of settings it belongs to.
#[derive(Debug, Clone, PartialEq, Eq, Reflect)]
pub enum SettingsType {
    /// These are settings that are saved in the os user's configuration directory. \
    /// These settings are global to the user and are not tied to a specific project. \
    /// Settings are along the lines of hotkeys etc.
    Global,
    /// Workspace preferences use the global preferences by default. End users can modify them, customizing their layout, theming and hotkeys. \
    /// The file is created when the user applies changes to their workspace preferences within the editor. \
    /// Workspace preferences can be shared between multiple projects and are not isolated to project folders.*
    Workspace,
    /// Project preference overrides are empty and stored within the project settings. \
    ///  When a project overrides a global/workspace preference, it is no longer possible to change them. \
    ///  In order to modify the preference, users must modify the project settings instead.
    /// There are two states that overrides can be in:
    /// - Inheriting - No override is set. Users can freely change the preference. Users can use what they have set within the global/workspace preferences.
    /// - Modified - When an override has been set, users can no longer change the preference without modifying the project settings. You can switch between inheriting and modified at any time without consequence.
    Project,
}
```

## List merging strategy
```rs
#[derive(Debug, Clone, Reflect, Default)]
/// Annotation for a type to show how to merge lists when loading settings.
/// if not set, the default is to replace the existing list.
pub enum MergeStrategy {
    #[default]
    /// When Mergeing the list, the new list will replace the existing list.
    Replace,
    /// When Mergeing the list, the new list will be appended to the existing list.
    Append,
}
```

## Tags
```rs
#[derive(Debug, Clone, Reflect)]
/// Annotation for a type to add tags to the settings. these tags can be used to filter settings in the editor.
pub struct SettingsTags(pub Vec<&'static str>);
```


hcabel:
This is super exiting, I really hope we can use this soon 🤞 Let me know if you need help :)

crates/bevy_editor_settings/src/local_prefs.rs
```diff
@@ -0,0 +1,30 @@
...
+impl Default for Build {
+    fn default() -> Self {
+        Self {
+            debug: "cargo build".to_string(),
+            release: "cargo build --release".to_string(),
+            run_debug: "cargo run".to_string(),
+            run_release: "cargo run --release".to_string(),
+        }
+    }
+}
```
MiniaczQ, line 30: Shouldn't we just generalize this as "profiles" or "tasks"?
Local builds can be different than CI builds, also we could use profiles for various tests

Since we're hardcoding this right now anyways, we might as well do
```rs
Profile {
  name: "Debug Build".to_string(),
  command: "cargo build".to_string(),
  description: None,
}
```
and store it in a file

n1ght-hunter: sorry meant to reply to this. that specific code was put together in like two mins. but the idea is you can override the build commands in a toml file at the root of the project. so the editor know how to start the game no matter the build system.

n1ght-hunter: btw not against your changes but would you be able to explain why the editor needs to know how to build in the app in ci?

n1ght-hunter:
update on this pr i am currently working on rewriting the code to match closer to the bevy reflect serde directory structure
https://github.com/bevyengine/bevy/tree/main/crates/bevy_reflect/src/serde

crates/bevy_editor_settings/src/local_prefs.rs
MiniaczQ: Editor should provide basic CLI for automation, even if it's technically another app, it's within the scope of editor as a project

n1ght-hunter: i think i will remove all this for this pr anyways. it was kinda based on my original design i went away from. it should be super easy to add back in with some more thought

n1ght-hunter:
@hcabel  @MiniaczQ other then a code cleanup this pr should be pretty close to completion. so feel free to have a look  through the code and review.

PR #80, opened 2024-10-15
Pane Layout System

tim-blackbird:
- Added a pane system based on nested flex nodes
- Added a stupid simple `PaneRegistry` for the pane crates to hook into.
The dependency between `bevy_3d_viewport` and `bevy_panes_layout` has been inverted to make use of this.
- Adjusted some colors to match mockups from felinusfish
 on discord, instead of Blender's colors :v
- Reduced border radius (fite me)

This is lacking a way to add new panels -- other than splitting an existing one -- and a way to change a panel's type(e.g. Scene Tree, Viewport) at the moment.

No drag & drop yet either.

#### Todo
- Docs
- Appease clippy
- clip UI when the pane gets too small 

## Testing
Checkout the PR and `cargo run`

### Controls
I'd put these controls in a context menu, if we had one. Instead we have this:
- `Right click` a panel header to split it into two vertically. Hold `Left Shift` while clicking to split horizontally.
- `Middle click` a panel header to remove that panel.
- `Left click` and drag the regions between panels to resize the areas.

image


crates/bevy_pane_layout/src/lib.rs
```diff
@@ -15,125 +20,229 @@
...
+    spawn_pane(&mut commands, &theme, 0.6, "Properties").set_parent(sub_divider);
+
+    spawn_resize_handle(&mut commands, Divider::Horizontal).set_parent(divider);
+    spawn_pane(&mut commands, &theme, 0.8, "Viewport 3D").set_parent(divider);
+}
+
+#[derive(Component)]
+struct ResizeHandle;
+
+// fn add_dividers(
```
Atlas16A, line 183: Probably need to remove this commented out code unless its important?

alice-i-cecile (approved):

PR #82, opened 2024-10-16
Cleanup unnecessary `allow` attributes

tim-blackbird:
- Also yeet a commented line from #80

There's no code-quality label :(


alice-i-cecile:
I've cleaned up our labels :)

PR #92, opened 2024-10-18
Fix Pane Jitter

### LINKED ISSUE #91: Panes glitching / bugging on pane and window resize
When resizing a pane it sometimes glitches out and it seems quite laggy.
Also when fullscreening the editor nothing updates anymore and the program freezes.

https://github.com/user-attachments/assets/cc8f15f9-0d13-4004-96ba-94ffacefe62a

tim-blackbird:
- Making the position `Absolute` seems to Fix #91 :shrug: 
- Fix panic when resizing viewport pane down so the image node size becomes less than `1px`


hcabel:
Tested this fix, the jitter does indeed go away, but the UI being clipped off the window that I've mentioned in the UI is still there

PR #95, opened 2024-10-19
Add bevy wayland feature

Dokkae6949:
Adding this allows the editor UI to work better under `wayland` without weird `xwayland` rendering glitches / issues.

This partially fixes the issues in #91 since fullscreen windows now render correctly. The pane glitching is not fixed by this tho

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ unsafe_op_in_unsafe_fn = "warn"
 unused_qualifications = "warn"
 
 [workspace.dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy.git", rev = "015f2c69ca2a2c009fd04eadada282482deaf469" }
+bevy = { git = "https://github.com/bevyengine/bevy.git", rev = "015f2c69ca2a2c009fd04eadada282482deaf469", features = ["wayland"] }
 thiserror = "1"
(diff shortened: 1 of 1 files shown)


tim-blackbird:
If wayland support is ready for prime-time then we should enable it in bevy by default, and I think we should not enable it here until then

Dokkae6949:
Can I enable the feature just for myself locally somehow? Since xwayland (without this feature flag) bevy is completely borked and borderline unusable

alice-i-cecile:
I don't think that's particularly possible. I agree that this is annoying, but I'm also nervous to turn this on locally 🤔

Dokkae6949:
From what I've seen is that in wayland is already enabled by default on the dev branch. So it really doesn't make sense to not already enable it since it works better than xwayland

tim-blackbird:
It's not enabled yet, if it was it would be enabled here too because we are on bevy's main branch

alice-i-cecile:
Maybe this is a good opportunity to dogfood it? If users say this is better than not enabling it, I think that merging this is the right call.

tim-blackbird:
I suppose enabling it here would have a manageable impact. I checked and it seems to run fine for me on Wayland with the feature enabled, so I'm cool with enabling it here :)

PR #78, opened 2024-10-13
Asset browser V1

hcabel:
Image

# Features
1. Reflect the assets directory on disk
2. Auto sync if any change is made to the asset directory 
3. TopBar show the current path
4. Any segment of the TopBar path is clickable to go back to the desired directory
5. Folder assets can be clicked to step into them
6. I've made 2 icon in figma to get things started

I've decided to go this way because from my experience I've realize people tends to prefer asset browser that are synced with the disk.

# Testing
To test it, you can bind it to the editor UI like so:
1. Go to the `bevy_editor` crate
2. Add `bevy_asset_browser.workspace = true` to `Cargo.toml` 
3. Open `src/ui.rs`
4. Import `use bevy_asset_browser::{AssetBrowserNode, AssetBrowserPlugin, AssetBrowserSet};`
5. In the plugin build function: configure_sets `AssetBrowserSet` and add_plugins `AssetBrowserPlugin`
6. then add `parent.spawn(AssetBrowserNode);` to the UI

# Notes
- I don't know where the game assets will be stored, so for now I've target the assets folder in the root directory
- To sync the asset browser to any external changes, I'm using the folder's metadata's modified time. I've stored the last modified value in a resources but I'm not sure it belong (since it has no purpose externally), is that ok?
- Please check for typos or weird phrasing, I'm not a native English speaker. (you've probably already noticed that 😆) trying my best
- I'm not using `bevy_pane_layout`, I haven't seen any other crate used it yet, so I'm not sure how to use it or if it's even finished

# Disclaimer
I'm here to help, if anything's ain't to your liking I'll be happy to change it. (My concern atm is that it might what you've envisioned)
I'm not a Rust nor a Bevy expert, I'll do my research but if I didn't use a rust/bevy feature, it's probably that I didn't knew it existed and I'll be happy to rework my code accordingly.

PR #96, opened 2024-10-19
Scroll box widget

hcabel:
⚠️ #78 need to be merged first

Adding a `ScrollBox` widget
scroll_bar_demo
(also the bar disappear if there is no content clipped) 

At the moment, the `ScrollBar` is being recalculated every frame.
It's far from being efficient, but I don't know how to react to UI being resized
I recon we could refresh on scroll and windows resizing event, but there's still cases were it might break (e.g pane_layout resize)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ bevy_command_palette = { path = "bevy_widgets/bevy_command_palette" }
 bevy_context_menu = { path = "bevy_widgets/bevy_context_menu" }
 bevy_i-cant-believe-its-not-bsn = { path = "bevy_widgets/bevy_i-cant-believe-its-not-bsn" }
 bevy_menu_bar = { path = "bevy_widgets/bevy_menu_bar" }
+bevy_scroll_box = { path = "bevy_widgets/bevy_scroll_box" }
 bevy_footer_bar = { path = "bevy_widgets/bevy_footer_bar" }
 bevy_toolbar = { path = "bevy_widgets/bevy_toolbar" }
 bevy_tooltips = { path = "bevy_widgets/bevy_tooltips" }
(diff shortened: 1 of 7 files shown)


hcabel:
Reopening, with a cleaned history, but it's the same

alice-i-cecile (dismissed):
Solid stuff; very pleased to see this sort of foundational work. Let me know when you've resolved merge conflicts.

tim-blackbird:
Bevy recently gained some built-in support for scrolling. See [`Overflow::scroll_y()`](https://dev-docs.bevyengine.org/bevy/ui/struct.Overflow.html#method.scroll_y) and [`ScrollPosition`](https://dev-docs.bevyengine.org/bevy/ui/struct.ScrollPosition.html). We can make use of that here :)

hcabel:
Actually this is not ready yet :D 

While changing the spawn method to allow creating child into the `ScrollBoxNode` (content) (🤔 I should rename that)
I stumble upon couple of issues:
1- it scroll no matter where the cursor is. And adding an Active tag on over/out is tricky because the ScrollBox is beneath other UI element
2- Doesn't support multi `ScrollBar`, (crash)

So I'm gonna keep improving this when I got more time and let you know when it's done :)

hcabel:
# Features
- no more crashes 🤞 
- Work in X/Y/Both axis
- Use Overflow and ScrollPosition
- Scroll bar doesn't disappear anymore if the size of the content is too small (query conflict) 
- Shift key + scroll, scroll the ScrollBox horizontally

# Notes
The scroll bar handle aren't yet draggable
Scroll bar handle are still recalculated every frame (`Update`)

hcabel:
@Atlas16A You can merge it now, ain't perfect but the most important is there
Allow people to create scrollable content without having to do it themselves

rewin123:
> @Atlas16A I recon you can merge it now, ain't perfect but the most important is there Allow people to create scrollable content without having to do it themselves

I really want this

PR #98, opened 2024-10-19
Adjust styles of panes

Atlas16A:
image

Adjusts the styles of the panes, change clearcolor of viewport camera to utilize theme.

### DIFF
--- a/bevy_editor_panes/bevy_2d_viewport/Cargo.toml
+++ b/bevy_editor_panes/bevy_2d_viewport/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 bevy.workspace = true
 bevy_pane_layout.workspace = true
 bevy_editor_camera.workspace = true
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

tim-blackbird:
We already look too much like Blender, copying their colors too is a bad idea :sweat_smile:

Atlas16A:
> We already look too much like Blender, copying their colors too is a bad idea :sweat_smile: 

I was mainly doing it to get the color contrast balance, previously my eyes kinda were hurting are the inverted color balance

tim-blackbird:
That's fair, I have no problem with changing the colors at all, but looking like Blender is a bad look for us.
We should find something nice that sets us apart.

Atlas16A:
That is the plan, I'm just not worried about looking like blender for how early we are rn.

PR #105, opened 2024-10-20
Vendor `bevy_editor_cam`

tim-blackbird:
- The first commit is a direct copy of relevant files from https://github.com/aevyrie/bevy_editor_cam.
  See the later commits for the changes I've made

### Notes
- SSAO in the `cad` example is really jittery, not sure what's different from the `ssao` bevy example which works fine
- We don't have `big_space` so the associated example was dropped


alice-i-cecile:
Let's swap to basic meshes for now please. I'd like sample assets but we should be more deliberate about them.

alice-i-cecile (approved):
Strongly in favor of using this.No issues detected with the initial port :) Thanks for swapping to simple assets.

alice-i-cecile:
The `bevy_mod_picking` reference failing CI should be swapped to plain `bevy_picking` now :D

PR #109, opened 2024-10-21
Fix Asset browser bugs

### LINKED ISSUE #108: Crash when splitting the AssetBrowser UI
# Reproduce steps
- `cargo run`
- right click on the "Asset Browser" window
- Split vertically or horizontally

hcabel:
### Fix AssetBrowser split pane crash (#108)
AssetBrowser pane was only supporting 1 instance at a time, now support N pane.
**Isn't a proper Multi pane support as every pane are in sync** but it doesn't crash anymore 

### Fix AssetBrowser height overflowing
Viewport pane scale + AssetBrowser pane scale, was over 1.0 which was "overflowing" at the bottom
(no issue to track this, @Atlas16A mentioned it on Discord)


tim-blackbird:
I clicked around a bit and managed to crash on `path_list_query.single()` at `bevy_asset_browser/src/lib.rs:290:66` :)

Atlas16A:
I think this is a good sign that we need to improve the pane layout implementation so that these issues aren't possible, it shouldn't be the responsibility of the pane to make sure it obeys multipane.

hcabel:
> I clicked around a bit and managed to crash on `path_list_query.single()` at `bevy_asset_browser/src/lib.rs:290:66` :)

Good catch! Is there a `.single()` that return an Error instead of panicking?

alice-i-cecile:
> > I clicked around a bit and managed to crash on `path_list_query.single()` at `bevy_asset_browser/src/lib.rs:290:66` :)
> 
> Good catch! Is there a `.single()` that return an Error instead of panicking?

`Query.get_single` :)

PR #110, opened 2024-10-21
3D Viewport

tim-blackbird:
Work in progress


Atlas16A (dismissed):

alice-i-cecile:
> Implemented render-target picking pass-through. Needed for bevy_editor_cam but also useful generally. That code should be moved somewhere else, later. Perhaps even upstreamed to bevy_picking

Yes please to upstreaming.

PR #118, opened 2024-10-24
Reduce footer height

tim-blackbird:
- Reduced footer height to `20px`. This is the height used by both VS Code and Blender
- Cleaned up some unused style parameters

### Showcase
Before
image

After
image


hcabel:
Now that you've point it out, your right it was huge
I agree and tested this change 👍

Atlas16A (approved):
Yea this is fine for now, I had originally set it up with plans to use selection context to give extra shortcut state info like blender does, but since itll be awhile before we can do that makes sense to simify it now.

PR #124, opened 2024-10-24
Add menu item for loading GLTF files

tim-blackbird:
I originally wanted to integrate this into the asset browser, but that is tricky and needs more design work.
So, instead you can:
* Press `Ctrl + L` to `L`oad a file via the native file-picker, or
* Load a file by dropping it on the window
  Note: Not supported on wayland: https://github.com/rust-windowing/winit/issues/1881

if a file is loaded then the first scene in that file will be spawned.

I've replaced the default cube with a plane to it doesn't overlap with the loaded scene, and shadows are now enabled on the directional light.

### Showcase
image

### Future Work
- This is temporary. We shouldn't load GTLF files directly, we should import them and then we can load them from the asset browser.
- We need a way to add 'actions' to asset types in the asset browser

### DIFF
--- a/crates/bevy_editor/Cargo.toml
+++ b/crates/bevy_editor/Cargo.toml
@@ -11,6 +11,8 @@ bevy_footer_bar.workspace = true
 bevy_context_menu.workspace = true
 bevy_editor_styles.workspace = true
 
+rfd.workspace = true
+
 # Panes
 bevy_3d_viewport.workspace = true
 bevy_2d_viewport.workspace = true
--- a/crates/bevy_editor/src/load_gltf.rs
+++ b/crates/bevy_editor/src/load_gltf.rs
@@ -1,0 +1,73 @@
+//! This is a temporary solution for loading GLTF scenes into the editor.
+
+use bevy::{
+    prelude::*,
+    tasks::{block_on, futures_lite::future, AsyncComputeTaskPool, Task},
+};
+use rfd::{AsyncFileDialog, FileHandle};
+
+pub(crate) struct LoadGltfPlugin;
+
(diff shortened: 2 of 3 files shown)


Atlas16A (changes requested):
I'd merge this after its moved to a more appropriate crate, perhaps the main bevy_editor crate? It could even live as an example under it. 

To avoid circular deps, put a marker component on it and import that component into main (at least for now)

n1ght-hunter:
i would recommend bevy_editor_utils. we are going to need more things like this anyways.

hcabel:
Yeah what's been holding my hand on the AssetBrowser is **where is the project related stuff is going to go?**
Then I'm more than happy to work something out to drag and drop asset in the viewport to spawn them

Atlas16A:
> Yeah what's been holding my hand on the AssetBrowser is **where is the project related stuff is going to go?**
> Then I'm more than happy to work something out to drag and drop asset in the viewport to spawn them
>  

Make an entirely separate bevy rust project directory outside the editor repo, that's your target for the asset browser, aka "target any arbitrary path and search for a assets folder"

tim-blackbird:
The code has been moved to `bevy_editor`. 
I've removed the `Load GLTF` button from the menu bar. Instead, you can now:
- Press `Ctrl + L` to `L`oad a file via the native file-picker, or
- Load a file by dropping it on the window
  Note: Not supported on wayland: https://github.com/rust-windowing/winit/issues/1881

> i would recommend bevy_editor_utils. we are going to need more things like this anyways.

I'd rather not add a `utils` crate if we can help it. The code in this PR is entirely temporary, it will be reworked soon enough, so lets not add a crate just for it.

PR #126, opened 2024-10-24
Rework theme crate

Atlas16A:
This should improve code clarity some.


n1ght-hunter (approved):
look good to me.
 i do think that copying the MUI create theme function could be a useful layout style i.e. having a palette struct etc

Atlas16A:
> look good to me.
>  i do think that copying the MUI create theme function could be a useful layout style i.e. having a palette struct etc

What's MUI?

n1ght-hunter:
> > look good to me.
> > i do think that copying the MUI create theme function could be a useful layout style i.e. having a palette struct etc
> 
> What's MUI?

Sorry Material UI.
https://bareynol.github.io/mui-theme-creator/
i think this ^ shows cases what i mean. you only need to pass in two colors and then it uses those to build whole ui styling from.
anyways this can be merged. just a thought because i have worked with MUI a lot and IMO there theming is one of the best i have used

PR #131, opened 2024-10-26
Change Z axis color

Jondolf:
The current Z axis color is way too saturated and hard on the eyes, and its perceived lightness is much lower than lightness of the red axis.

kuva

In most other tools and editor, like Godot and Blender, the Z axis uses a lighter blue.

kuva

kuva

## Solution

Change the axis color to a lighter blue that is close to the color used by Godot and Blender.

kuva


Atlas16A (commented):
Can you add the axis colors to the style crate and change them to use oklch?

(I can merge if you don't wanna rn)

Jondolf:
Yep I can do that

Atlas16A (approved):

Jondolf:
@Atlas16A Done, verified that colors are still correct too

kuva

PR #132, opened 2024-10-26
Infinite scaling grid

tim-blackbird:
The infinite grids now also scale infinitely when you zoom out!*
<sub>*Until they disintegrate due to floating point error</sub>

I've reworked the meaning of the grids `scale` value too. It now means 'world units per grid square' at the lowest scaling.
`fadeout_distance` is now also relative to the scaling.

### Future Work
- The transition is currently instantaneous and a bit jarring, this could be smoothed out

### DIFF
--- a/crates/bevy_infinite_grid/src/render/grid.wgsl
+++ b/crates/bevy_infinite_grid/src/render/grid.wgsl
@@ -95,14 +95,47 @@ fn fragment(in: VertexOutput) -> FragmentOutput {
 
     out.depth = clip_depth;
 
-    let scale = grid_settings.scale;
-    let coord = plane_coords * scale; // use the scale variable to set the distance between the lines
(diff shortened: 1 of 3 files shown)


crates/bevy_infinite_grid/src/render/grid.wgsl
```diff
@@ -95,14 +95,47 @@ fn fragment(in: VertexOutput) -> FragmentOutput {
...
+
+
+    // Orthographic scaling
+
+    // The height of the view in world units
+    let view_area_height = 2. / view.projection[1][1];
+
+    // Who knows what it means?
+    let cool_magic_number = 300.;
+    let size = view_area_height / cool_magic_number;
```
tim-blackbird, line 116: I've tried my best to document the new shader code but my brain is cooked on this one for now

PR #133, opened 2024-10-26
Pane select button 

Atlas16A:
Does nothing atm
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@ Low-friction experiments for the bevy_editor.
 
 Read about what we're building in the [Design Book](https://bevyengine.github.io/bevy_editor_prototypes/)!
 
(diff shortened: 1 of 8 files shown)


alice-i-cecile:
Can you say a bit more about what functionality this is supposed to have? Ideally with a comment in the code. I'm fine to merge WIP stuff like this, but it's easy for it to get confusing if there's no note on intent.

Atlas16A:
Sure I'll make a commit in just a bit, simple awnser is its the button to click to change a pane from say, the 2d viewport, into a properties pane, etc 

I just want to get it in there so its already in place when a drop down menu is available to implement on it (might do that unless Tim beats me to it lol)

alice-i-cecile (approved):

PR #135, opened 2024-10-27
Launcher and Projects

hcabel:
# Testing
- Go to `templates/blank_project` 
- in `Cargol.toml` change `bevy_editor` dependency for:
`bevy_editor = { git = "https://github.com/hcabel/bevy_editor_prototypes.git", branch = "editor_project" }`
(This is because the EditorPlugin doesn't exist on the main branch yet)
- `cargo run`

# Crate editor project
This is 1 of 2 idea I had to work on a per-project basis with the editor.
The idea here is for every project to be they're own crate that depend on the bevy editor plugin (which will be deactivated on packed build)
to run a project you'll only have to do `cargo run`
## Benefits
- Normal Bevy project
- Can easy add plugin and custom sources per project
- The default source point to the project `assets` folder
- Doesn't need recompiling when switching project
## Cons
- Haven't figure out how to set the name yet. would love to access to the crate name, otherwise will probably have a config somewhere
- Might be scary for non-developer users 

# Project file
This idea can be seen in the `project` branch of my fork
This isn't a crate but rather a project file at the root of the project folder, which the editor can load to get relevant information about the project (`path`, `name`, ...)
(I'm mentioning it because I've explored this, but I don't think what we want)
BevyProjectFile

### More ideas? happy to hear them


hcabel:
Oupsy I wanted to make this a Draft :D 

## Todos:
- We probably need a `editor_launcher` crate, which will be the default workspace crate.
This new crate will show a launcher UI with all the project you've already runner on your computer and an option to create new ones.
(For this feature we can have a config file in the appdata with all the project path already ran)
- We also need a better way for us to run a default project that doesn't depend on the git repo main branch. but from the local files.
At the moment you'll have to update that manually in the blank project.
(An easy fix for that would be to create a new template (could even make this one the default workspace project))

hcabel:
I've converted the `alien_cake_addict` project that we had to allow to be ran from the editor.

I had to create a new callback that is called when running the project without the editor (`cargo run -- -game`)
The good thing about this is that we don't have to worry about reverting the state of the game ^^

hcabel:
# Overview
## The launcher
The launcher is now the default app of the workspace.
### It's a vey ugly app that allow you to:
- Manage the projects on your machine.
It uses a cache file (`%localappdata%/Bevy Editor/projects.ron`/`~/.cache/Bevy Editor/project.rons`/`~/Library/Caches/projects.ron`)
This allow projects to be independent and to keep track of them even when not launched through the launcher.
Currently only the `path` and `last_opened` data are store

- Create new projects
Copies one of the templates (currently `blank_project`) to the user's selected destination folder.
The name of last folder will be use as the name of the project (currently useless)

- Run projects
Project can be runed by clicking on them, this ultimately just call `cargo run` at the root of the project.
If the project doesn't exist/isn't valid anymore it will be deleted from the project cache file (until it's runed successfully once)

### We have templates
Template have a lot of purposes, the main one being creating new project, but also it will be pretty handy for development to test how to integrate game specific code.
It also allow us to have a starting point for any future documentation: walk through, tutorial, ... 

#### As of right now 2 templates are available:
`blank_project`: Empty project, only run the editor when runed
`getting_started`: This is the old `alien_cake_addict` project, though as we don't support game code yet, it does nothing. But the goal is to have a very small project that people can use to explore the editor's feature.

# Concerns/Thoughts
- In the future will need to add the version of the project, in the `ProjectInfo`
- Creating a project copy a local folder. If this one is altered/missing it can lead to undefined behaviour
- We should modify templates once copied to reflect project infos (e.g. change crate name in `Cargo.toml` to match project name)
- New App. This is just to get me started, I needed something that every project could call to compile and run the editor.
My idea was that this App would work the same way more traditional engine would.
Where you call it, and it handle everything for you based on the config values you've set in the Editor UI. (stored in file, once this feature is ready)
I've made a quick config example as it would still be configurable through code, adding plugins, linking game code, ...
Running your project using `cargo run` will run your project in editor mode, and if you do `cargo run -- -game` it will run in a standalone mode. 
The `-game` flag tell the App what to do, and what Plugin should/shouldn't be used

alice-i-cecile:
We've [discussed](https://discord.com/channels/691052431525675048/1301174282893656167) this architecture pretty extensively, and I think a launcher is the right choice, at least for now. Let's merge this now and refine.

PR #138, opened 2024-10-30
Add basic glossary

doup:
Add a basic Glossary page. Initially based on Blender terms, changed after some discussion.

Once we decide in some basic terms we can update the whole prototype docs in a follow-up PR.

### DIFF
--- a/design-book/src/SUMMARY.md
+++ b/design-book/src/SUMMARY.md
@@ -2,6 +2,7 @@
 
 - [Vision](./vision.md)
 - [Roadmap](./roadmap.md)
+- [Glossary](./glossary.md)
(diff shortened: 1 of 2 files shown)


Jondolf (approved):
These terms seem adequately clear to me, and I think it's good to have shared terminology to make sure everyone is on the same page. We can add more terms as they pop up.

MiniaczQ:
I'd love to see the keywords be **bold** when used in the descriptions.

Also question, would we want `mermaid` diagrams to show relations between glossary items?

alice-i-cecile:
Restarted CI: hopefully this works now.

PR #143, opened 2024-11-03
Asset browser improvements (rewrite)

hcabel:
While trying to improve the Asset Browser crate I've realize that I didn't do a very good job and that this was hard to scale.
So I've rewrite as it was (- the hover animation) and here are the new features:

## new feature
- New context menu:
  - on AssetBrowser background: **Create new folder**
  - on Folders element: **Delete** (No confirm popup atm, be careful)
- Context menu improved:
  - Only 1 context menu can be opened at a time
  - When overlapping the context menu with the highest initiator zindex is the one that get's displayed
image
image


alice-i-cecile:
Nice!

mockersf:
You can't use methods from `std::fs` like `create_dir_all` and `remove_dir_all`. 

The `AssetWriter` exposes a method to delete a directory: https://docs.rs/bevy/0.15.0-rc.2/bevy/asset/io/trait.AssetWriter.html#tymethod.remove_directory so you should use this one instead.

It doesn't have yet a method to create a directory, you should first add it to Bevy

PR #145, opened 2024-11-04
Scene ~~tree~~ list

JMS55:
Only a scene list now, not actually a tree. But we can build on this more later.

Also does not support multiple copies of the pane atm.


bevy_editor_panes/bevy_scene_tree/src/lib.rs
```diff
@@ -1,17 +1,158 @@
...
+                  mut selected_entity: ResMut<SelectedEntity>| {
+                if selected_entity.0 == Some(scene_entity) {
+                    selected_entity.0 = None;
+                } else {
+                    selected_entity.0 = Some(scene_entity);
+                }
+                trigger.propagate(false);
+            };
+
+        if let Ok(HasSceneTreeRow(row_entity)) = has_scene_tree_row.get(scene_entity) {
```
JMS55, line 115: Should've used Option<HasSceneTreeRow> here instead.

tim-blackbird (approved):

PR #146, opened 2024-11-04
Linux launcher fixes

hcabel:
- Fix Linux project not running
- Fix cache folder not being created

Haven't fixed the crashes reported by @Atlas16A, not sure if that's something due to my PR  
if it crash for you try using `WGPU_BACKEND=vulkan cargo run`


doup (approved):
I confirm that this fixes the issue in macOS. 👌

PR #147, opened 2024-11-10
Adds an example to the launcher crate for developer sanity

Atlas16A:
Simple, adds an example so we can run to see changes without messing with cargo and separate project files just to see a change on editor side.

Run via `cargo run --example simple_editor`


hcabel:
What I've been doing was going into the blank template and changing the `Cargo.toml` to use my local bevy_editor crate (and all the others by extent)  `path: "../../crates/bevy_editor"`
This also allow to have an asset folder if you're messing with asset management

Though I'm onboard with this change :+1:

PR #148, opened 2024-11-10
Import ISSUE TEMPLATE

LoweredgamesDev:
- import into main Bevy project


alice-i-cecile:
I think this is a good starting point :)

PR #151, opened 2024-11-21
Basic assets preview and preview caching

443eb9:
Preview assets that marked `AssetPreview`. (might change in the future)

For 2d sprites, this should be a simple sprite 2d.
For meshes, we will render this asset at a low resolution, store the result and display it once needed.
For fonts, render some characters with that font.

TODOs
- [x] Generate preview image for 3d models.
- [x] Cache preview image.
- [ ] Regenerate when asset changes, including change during hot reloading.
- [ ] Automatically adjust the distance between camera and model according to model scale.


443eb9:
> @443eb9 just wanted to check if you're still working on this PR?

Nope, I'm quite busy these days. :(

Sieluna:
@jbuehler23 I'd like to help with this >w<

PR #160, opened 2024-11-30
Simplify context menu bugfix + option border radius adjustment

tim-blackbird:
The context menu bugfix from https://github.com/bevyengine/bevy_editor_prototypes/pull/143 was rather complicated. Here it is replaced with a `trigger.propagate(false);` call to prevent the right click event from bubbling up further when a context menu is opened.

I've also separated the context-menu option border-radius from general buttons and adjusted it to look right.

image


BUGO07:
looks good :+1:

PR #161, opened 2024-11-30
Add `bevy_editor_core` crate

tim-blackbird:
We need a crate to house things all other crates depend on.

- Added `bevy_editor_core` crate
- Moved `SelectedEntity` into `bevy_editor_core`


n1ght-hunter (approved):
This looks good to me

PR #163, opened 2024-12-01
add click observers and conditions for top bar items (even logo)

BUGO07:
I've tested it and it works, just haven't added any actions yet


BUGO07:
i mean't to add the second commit to a new pr but accidentally added it to this one. still check the changes out they're useful

PR #164, opened 2024-12-06
small improvements and a "bugfix" | context menus

BUGO07:
when opening the context menu, if you were holding down the right click and then hovered over a context menu option, when you'd release the right click it would execute that context menu action, that's been fixed by making the context menu spawn trigger be `Pointer<Up>` instead of `Pointer<Down>`, and also made the context menu options to trigger only on left click (primary button)

### DIFF
--- a/bevy_widgets/bevy_context_menu/src/lib.rs
+++ b/bevy_widgets/bevy_context_menu/src/lib.rs
@@ -17,7 +17,7 @@ impl Plugin for ContextMenuPlugin {
 }
 
 fn on_secondary_button_down_entity_with_context_menu(
-    mut trigger: Trigger<Pointer<Down>>,
+    mut trigger: Trigger<Pointer<Up>>,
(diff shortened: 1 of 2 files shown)


BUGO07:
this pr also makes it so when a context menu is open and you try to open another, it closes the currently open one and opens the new one. previously you would need to right click twice for this (once to close the current one and the second time to open the new one)

tim-blackbird:
All of this was designed on purpose by me. Context menus generally behave this way.
I can't confirm it works this way on MacOS, but it does for Linux KDE and VS Code

BUGO07:
I didn't know. I think it can be a bit annoying when you're moving the cursor while you right click and it does an action, especially since there's delete buttons in the context menus

BUGO07:
> So we need a way to decide how the context menu should behave on different platforms.

I think we can do that with preferences after we get there

PR #167, opened 2025-01-11
Template-based scene tree + inspector

JMS55:
TODO: Not sure how to add hover styling to rows yet. Going to remove it for now, we can figure out how to add it back later.

Things could maybe be a little more declarative to combine the setup pane and update rows into one template, but I'm fairly happy with how it's setup currently.

### DIFF
--- a/bevy_editor_panes/bevy_scene_tree/src/lib.rs
+++ b/bevy_editor_panes/bevy_scene_tree/src/lib.rs
@@ -50,8 +50,7 @@ fn update_scene_tree(
     for scene_tree in &scene_trees {
         let tree_rows: Template = scene_entities
             .iter()
-            .map(|(entity, name)| scene_tree_row_for_entity(entity, name, &selected_entity))
-            .flatten()
+            .flat_map(|(entity, name)| scene_tree_row_for_entity(entity, name, &selected_entity))
             .collect();
 
         commands.entity(scene_tree).build_children(tree_rows);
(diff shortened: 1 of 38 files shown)


bevy_editor_panes/bevy_scene_tree/src/lib.rs
```diff
@@ -57,122 +41,53 @@ fn setup_pane(pane: In<PaneStructure>, mut commands: Commands) {
...
-impl Component for HasSceneTreeRow {
-    const STORAGE_TYPE: StorageType = StorageType::Table;
-
-    fn register_component_hooks(hooks: &mut ComponentHooks) {
-        hooks.on_remove(|mut world, entity, _| {
-            let row_entity = world.entity(entity).get::<Self>().unwrap().0;
-            world.commands().entity(row_entity).despawn_recursive();
-        });
+    template! {
+        {entity}: (
```
NthTensor, line 76: It hadn't occurred to me that you would be able to use entities as anchors like this. Pleased to see that's not too complex.

NthTensor (approved):
Sweet! I notice you didn't end up needing the `b!()` macro or conditional stuff too much. Was that something you avoided due to the constraints of the macro, or just not something that was needed?

bevy_editor_panes/bevy_scene_tree/src/lib.rs
JMS55: Yeah apparently entity's impl whatever trait you use for naming, which is great because otherwise the list order / hover state / etc would break when renaming an entity, if I went off of Name.

JMS55:
Tbh I never tried it. I ended up writing functions returning Template, using a branch to return one of two different templates, and then splicing it into a larger template. This was just the natural way for me to write things, I didn't actually consider using conditional stuff.

But also there's not much state in these widgets so far, they're pretty much all view-only. I might end up needing them later, no idea.

PR #171, opened 2025-01-19
Add BSN prototype

villor:
## BSN Prototype
This PR adds a BSN prototype. Supporting many (but **not** all) of the concepts outlined in [Cart's proposal](https://github.com/bevyengine/bevy/discussions/14437).

My hope is that this will serve as a good-enough foundation to allow progress on scene editor prototyping!

To keep this on a prototype level, everything is placed in a single crate called `bevy_bsn`, with everything `pub` exposed at the top level.

### 📦 Features
- `Construct`/`Patch` - the foundational building blocks
- `EntityPatch`/`Scene` - The underlying types for the `bsn!` macro, allowing static scene construction.
- `DynamicPatch`/`DynamicScene` - Reflection-powered scene patching/construction.
  - I could not figure out the intended implementation here (something utilizing `Downcast`), so I went with a workaround (`ReflectPatch`) for now.
- `bsn!`-macro
  - Mostly syntax equivalent to the proposal
  - Inheritance is supported, but _probably not_ in the exact intended way.
- `.bsn`-asset
  - A single tree of entities - no BSN Sets or similar features
  - Supports loading files as a `Bsn` asset, which is a non-type aware AST basically.
  - `Bsn` can be turned into a `DynamicScene` using reflection, which can then be spawned.
  - `Bsn` can be converted into a non-pretty string (`Bsn::to_bsn_string`) representation for saving.
  - Reflection is done using `BsnReflector`, API is very bare-bones and needs design thinking, but I figure we will come up with good solutions when implementing the first BSN editor.
  - No inheritance support
- Utilities/workarounds for a better BSN-experience
  - BSN helper functions like `px`, `px_all`, `rgb8`, `rgba8`.
  - Newtypes with `Construct`-implementations for types that are currently `Default + Clone` upstream: `ConstructHandle`, `ConstructEntity`, `ConstructTextFont`
    - We can just add more of these as needed for prototyping, until `Construct` is available upstream.
  - Type registrations to allow construct reflection of a lot of upstream blanket-implemented types.
- A `BsnPlugin` adding support for `.bsn`-assets and construct reflection.

### 📺 Examples
BSN Macro showcase:
```sh
cargo run --example bsn_macro--package=bevy_bsn
```

BSN Asset loading and spawning:
```sh
cargo run --example bsn_asset --package=bevy_bsn
```

Roundtrip of loading, editing, and saving a BSN Asset:
```sh
cargo run --example bsn_edit --package=bevy_bsn
```

#### 1. **Inheritance**
Inheritance is not currently supported for `.bsn`-assets.

----------------------
#### 2. **Required components `construct`**
The `#[require(construct(Type))]`-addon for required components is not supported as that would(?) require upstream changes.

----------------------
#### 3. **Function reflection**
`FunctionRegistry` not supported for reflection, meaning bsn helpers are not available in bsn assets.

----------------------
#### 4. **BSN Asset Imports**
Type paths are resolved using `short_path`, which is not ideal. The proposal includes support for `use`-declarations, which may be a good idea to implement and keep in mind for the scene editor.

----------------------
#### 5. **Constants**
Constants are not supported when reflecting BSN assets. This means we can't use things like predefined color, e.g `css::FIRE_BRICK`.

One idea is to add our own `ConstantRegistry` for now, which could also be upstreamed to `bevy_reflect` at some point.

----------------------
#### 6. **Into reflection**
Reflection of BSN-assets does not currently handle `.into()` like the macro does.

Any ideas?

----------------------
#### 7. **Reflection APIs are in dire need of some love**
`BsnReflector` has no thought put into its API. It needs more work to power scene editing in a nice way.

The code isn't very clean either. Just a bunch of `reflect_x` functions which may work better as traits or something.

We may also be able to abstract it somehow to allow using a type registry from somewhere else, like BRP, which would enable editing scene files without statically linking the editor and game.

----------------------
#### 8. **More `ReflectConstruct` registrations!**
The list of registrations of `ReflectConstruct` for upstream blanket-Construct components is a bit lacking. It needs to be extended for any components that we may want to use in `.bsn`-assets.

See: `bevy_bsn::construct_reflect::register_reflect_construct`

### 💎 Bonus example powered by `i-cant-believe-its-not-bsn`
I have included a bonus example: `bsn_icbinbsn_hot`, which showcases non-destructive hot reloading of a `.bsn`-file powered by `Fragment`/`Receipt`. May be useful for some scene editor stuff.

```sh
cargo run --example bsn_icbinbsn_hot --package=bevy_bsn --features=bevy/file_watcher
```

Note that I made a small addition (`BoxedBundle::new`) to `bevy_i-cant-believe-its-not-bsn` to make this possible.

### 👥 Credits
- @cart - The brains behind BSN and prototype code to copy/look at
    - [Early prototype](https://github.com/cart/bevy/tree/bevy_scene2)
    - [Construct prototype](https://github.com/cart/bevy_construct_prototype)
- @shanecelis - The `Construct`/`Patch`-implementation was initially based on their implementation in [`bevy_asky`](https://github.com/shanecelis/bevy_asky/blob/main/src/construct.rs).


BUGO07:
I recommend changing the name to bevy_scene_notation or something like that because bevy_bsn has double bevy

alice-i-cecile:
I'm in favor of merging this in and experimenting with it here. @NthTensor we should probably roll back the i-cant-believe-its-not-bsn code from this crate in favor of consolidating on a single experiment to avoid *complete* chaos. This experiment is more featureful, so let's stick with it.

NthTensor:
Yeah, i'm happy to see things moving totally in favor of whatever needs to happen with my template macro. 

But this doesn't fully have the same incremental building that the template does right? Would it be possible to refactor jasmine's scene tree (https://github.com/bevyengine/bevy_editor_prototypes/pull/167) to use this?

villor:
> …we should probably roll back the i-cant-believe-its-not-bsn code from this crate in favor of consolidating on a single experiment to avoid _complete_ chaos. This experiment is more featureful, so let's stick with it.

I want to clarify that this PR does not really provide an attractive alternative to icbinbsn’s templates in its current state. It doesn’t have any strategies for incrementally updating a scene tree, which is the core benefit of `template` as it gives a very nice illusion of reactiveness/immediate mode.

It also doesn’t have equivalents for splices or named fragments.

My intention with this PR is primarily to provide the means for scene asset editing stories. I included the macro for completeness, but I wouldn’t really recommend using it over `template` right now.

However, I do think that there would be a benefit in consolidating the two. Merging the best of both to provide the ultimate tool to accelerate the editor UI development.

NthTensor:
> However, I do think that there would be a benefit in consolidating the two. Merging the best of both to provide the ultimate tool to accelerate the editor UI development.

Agreed, I think this is the path forward. We would could probably provide the features of the template macro to bsn very easily with minor syntax changes (optional names will be required but we could use the `Name` component theoretically).

I've not gotten a chance to assess this code, but I can't imagine adding basic receipt-based diffing incrementalism will be impossible. But I don't think it should block this PR, and I won't have time to work on anything like that for a few weeks probably.

villor:
> I recommend changing the name to bevy_scene_notation or something like that because bevy_bsn has double bevy

Can't say that I agree. The abbreviation BSN seems to be pretty established and recognized already (think JSON, nobody refers to it as JavaScript Object Syntax Notation). In general I'm pro verbosity, but only if it actually adds recognizability/readability. In this case I think `bevy_bsn` is clearer and more recognizable than `bevy_scene_notation` (even if there's a double bevy in there)

I'm happy to surrender and change it though if needed. No reason to bikeshed this for the prototype.

PR #193, opened 2025-06-02
Refine milestones and add a section on architecture

alice-i-cecile:
This PR is a cleanup and clarification pass, aimed at helping unblock project management and refocus developers here on achieving an MVP, rather than waiting on upstream features.

The main elements:

1. Add a mention of the design book to the README, in the hopes that people will immediately jump into that as a starting point.
2. Add an architecture.md section, which helps summarize the big project architecture questions that we've decided on, and lists out the open questions we still need research on.
3. Refines the milestones, shuffling work items around and focusing on the deliverable outcomes after each step.

If there's an information that's out of date, please correct me! If you disagree with my prioritization or architectural choices, now is the time to hash it out :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,14 @@
 # Bevy Editor Prototypes
 
 > [!WARNING]
-> This project is a prototype for a shippable Bevy Editor and is still in heavy development and testing. 
+> This project is a prototype for a shippable Bevy Editor and is still in heavy development and testing.
 > Everything you see is very much a work-in-progress. Read about what we're building in the [Design Book](https://bevyengine.github.io/bevy_editor_prototypes/)! or [Figma document].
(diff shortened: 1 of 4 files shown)


jbuehler23 (commented):
if no other comments/feedback i would get this merged ASAP so other devs can pick up

alice-i-cecile:
Ha, we need to fix the new clippy lints before we can merge anything. Bah!

jbuehler23:
I wouldn’t have thought a few readme changes would break that. Is it something with new versions maybe?

alice-i-cecile:
Nah, it's new changes to the Rust version in CI. We hit the same thing over in Bevy. Should be an easy fix; gimme a bit.

PR #195, opened 2025-06-03
Add OutlineGizmoPlugin and integrate outline gizmo functionality

jbuehler23:
First cut at getting an Outline Gizmo working - very rough, and highly WIP while learning the ins and outs of bevy and the repo :)

My first thoughts are that it seems _incredibly_ verbose to create a toggle which is updated by some "listener".

It would be nice if this block could be heavily simplified:

```
parent
        .spawn((
            Node {
                position_type: PositionType::Absolute,
                top: Val::Px(10.0),
                left: Val::Px(10.0),
                width: Val::Px(120.0),
                height: Val::Px(30.0),
                align_items: AlignItems::Center,
                justify_content: JustifyContent::Center,
                ..default()
            },
            BackgroundColor(Color::srgb(0.2, 0.2, 0.2)),
        ))
        .with_children(|parent| {
            parent.spawn((
                Text::new(""),
                GizmoToggleText,
            ));
        })
        .observe(|_trigger: Trigger<Pointer<Click>>, mut show_outlines: ResMut<ShowOutlines>| {
            show_outlines.0 = !show_outlines.0;
        });

// System to update the button text when ShowOutlines changes
fn update_gizmo_toggle_text(
    show_outlines: Res<ShowOutlines>,
    mut query: Query<&mut Text, With<GizmoToggleText>>,
) {
    if show_outlines.is_changed() {
        for mut text in &mut query {
            text.0 = if show_outlines.0 { "Hide Outlines".into() } else { "Show Outlines".into() };
        }
    }
}
```
  into something with a more simplified callback structure.

This is also a first step toward the Milestone Issue we have here: #197 

Also adds documentation outlined in #207 to give new devs help when trying compile/run changes locally

Screenshot of toggle button working on the editor locally :)

image

### DIFF
--- a/bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
+++ b/bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
@@ -20,14 +20,17 @@ struct GizmoToggleText;
 
 pub fn outline_gizmo_system(
     show: Res<ShowOutlines>,
-    selected: Query<&Transform, With<SelectedEntity>>,
+    query: Query<&Transform>,
+    selected_entity: Res<SelectedEntity>,
     mut gizmos: Gizmos,
 ) {
     if !show.0 {
         return;
     }
-    for transform in &selected {
-        gizmos.cuboid(*transform, Color::srgb(1.0, 1.0, 0.0));
+    if let Some(entity) = selected_entity.0 {
+        if let Ok(transform) = query.get(entity) {
+            gizmos.cuboid(*transform, Color::srgb(1.0, 0.0, 0.0));
+        }
     }
 }
 
--- a/crates/bevy_editor/src/lib.rs
+++ b/crates/bevy_editor/src/lib.rs
@@ -46,7 +46,7 @@ pub struct EditorPlugin;
 impl Plugin for EditorPlugin {
     fn build(&self, bevy_app: &mut BevyApp) {
         // Update/register this project to the editor project list
-        // project::update_project_info();
+        project::update_project_info();
         info!("Loading Bevy Editor");
         bevy_app
             .add_plugins((
--- a/crates/bevy_editor_core/src/lib.rs
+++ b/crates/bevy_editor_core/src/lib.rs
@@ -14,7 +14,7 @@ impl Plugin for EditorCorePlugin {
 }
 
 /// The currently selected entity in the scene.
-#[derive(Component, Resource, Default, Reflect)]
+#[derive(Resource, Default, Reflect)]
 #[reflect(Resource, Default)]
 pub struct SelectedEntity(pub Option<Entity>);
 


alice-i-cecile (commented):
I'm happy with the code here: verbose but clear. Toss in a screenshot to the PR description and I'll merge this for you.

bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
```diff
@@ -0,0 +1,77 @@
...
+impl Plugin for OutlineGizmoPlugin {
+    fn build(&self, app: &mut App) {
+        app.init_resource::<ShowOutlines>()
+            .add_systems(Startup, spawn_gizmo_toggle_ui)
+            .add_systems(Update, outline_gizmo_system)
+            .add_systems(Update, update_gizmo_toggle_text);
+    }
+}
+
+// Marker component for selected entities
```
Constanze3, line 13: There already is a SelectedEntity resource defined in bevy_editor_core, you should not define your own component.

bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
```diff
@@ -0,0 +1,77 @@
...
+pub fn spawn_gizmo_toggle_ui(mut commands: Commands) {
+    info!("Spawning Gizmo Toggle UI");
+    commands
+        .spawn((
+            Node {
+                position_type: PositionType::Absolute,
+                top: Val::Px(20.0),
+                right: Val::Px(20.0),
+                width: Val::Px(150.0),
+                height: Val::Px(15.0),
```
Constanze3, line 46: The height of this button is smaller than the text inside which is not very pleasing to look at.

bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
```diff
@@ -0,0 +1,77 @@
...
+) {
+    if !show.0 {
+        return;
+    }
+    for transform in &selected {
+        gizmos.cuboid(*transform, Color::srgb(1.0, 1.0, 0.0));
+    }
+}
+
+pub fn spawn_gizmo_toggle_ui(mut commands: Commands) {
```
Constanze3, line 37: This should be only a temporary solution, it would be better to have a menu to manage which gizmo types are visible just like other game engines.

crates/bevy_editor/src/lib.rs
```diff
@@ -46,8 +46,8 @@ pub struct EditorPlugin;
 impl Plugin for EditorPlugin {
     fn build(&self, bevy_app: &mut BevyApp) {
         // Update/register this project to the editor project list
-        project::update_project_info();
-
+        // project::update_project_info();
+        println!("Loading Bevy Editor");
```
Constanze3, line 50: You should rather use info!.

crates/bevy_editor/src/lib.rs
```diff
@@ -89,28 +90,7 @@ impl App {
...
-    ));
-
-    commands.spawn((
-        DirectionalLight {
-            shadows_enabled: true,
-            ..default()
-        },
-        Transform::default().looking_to(Vec3::NEG_ONE, Vec3::Y),
-    ));
+fn load_example_scene(mut commands: Commands, asset_server: Res<AssetServer>) {
```
Constanze3, line 93: This will crash the editor when it is not run as the example. This should be somehow integrated into the example code instead.

Constanze3 (changes requested):

crates/bevy_editor/src/lib.rs
jbuehler23: yes, i've removed this for now

crates/bevy_editor/src/lib.rs
jbuehler23: done

bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
jbuehler23: yes, i've fixed thisi by making the font smaller for now

bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
jbuehler23: done - i've added `bevy_editor_core` to the `cargo.toml` for `bevy_3d_viewport`

bevy_editor_panes/bevy_3d_viewport/src/outline_gizmo.rs
jbuehler23: image

jbuehler23:
@Constanze3 Thanks for the constructive review! This was so helpful, and I've made those changes you've recommended. I've also removed the example scene loading for now, until we come up with a more elegant solution.

crates/bevy_editor/src/lib.rs
```diff
@@ -46,8 +46,8 @@ pub struct EditorPlugin;
 impl Plugin for EditorPlugin {
     fn build(&self, bevy_app: &mut BevyApp) {
         // Update/register this project to the editor project list
-        project::update_project_info();
-
+        // project::update_project_info();
```
Constanze3, line 49: Was this commented out intentionally?

crates/bevy_editor_core/src/lib.rs
```diff
@@ -14,7 +14,7 @@ impl Plugin for EditorCorePlugin {
 }
 
 /// The currently selected entity in the scene.
-#[derive(Resource, Default, Reflect)]
+#[derive(Component, Resource, Default, Reflect)]
```
Constanze3, line 17: This is not how you should do this. This should not be a component.

Instead in your gizmo system you should have a query: Query\<&Transform\> and also grab selected_entity: Res\<SelectedEntity\> and then use query.get(selected_entity.0) to get the transform of the selected entity.

crates/bevy_editor/src/lib.rs
jbuehler23: oh strange, that must've been an accident on my end - apols

Constanze3 (approved):

PR #202, opened 2025-06-06
Fix Clone impl of bevy_proto_bsn's ReflectedValue

mmatvein:
Took bevy_proto_bsn to use and had trouble deserializing assets.

As far as I can tell, `PartialReflect::reflect_clone` always returns NotImplemented. Changed the `reflect_clone` call to `to_dynamic` and things seem to work. I had to make a similar change when updating our Blenvy version to support Bevy 0.16 after being on 0.15.

### DIFF
--- a/crates/bevy_proto_bsn/src/bsn_reflect.rs
+++ b/crates/bevy_proto_bsn/src/bsn_reflect.rs
@@ -207,11 +207,7 @@ impl Clone for ReflectedValue {
     fn clone(&self) -> Self {
         Self {
             type_id: self.type_id,
-            instance: self
-                .instance
-                .as_ref()
-                .reflect_clone()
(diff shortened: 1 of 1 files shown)


jbuehler23:
@villor if you can take a look please :)

villor (approved):
Looks great to me. `to_dynamic` is the natural migration path from the previous use of `clone_value`.

I guess maybe an even better way to do it after the recent reflect_clone changes would be to first try `reflect_clone` and then fall back to `to_dynamic` if not implemented. But I don't know if we have any reason to do so right now as most of the stuff we are cloning are dynamic objects already. Let's get this merged instead as it fixes a major bug.

Thank you!

mmatvein:
Formatting fixed!

PR #203, opened 2025-06-06
Add support for negative numbers in bevy_proto_bsn parsing

mmatvein:
Negative numbers are parsed as an Unary operation containing the negative sign (-) and the number itself, instead of just as the plain number, which resulted in negative numbers in proto_bsn assets being interpret as `BsnValue::UnknownExpr` instead of correctly as `BsnValue::Number`. This made it impossible to use negative numbers in proto_bsn assets.


villor (approved):
Nice fix for a pretty foundational bug. Good to see someone test it out properly! LGTM

PR #204, opened 2025-06-06
Add support for lists in bevy_proto_bsn parsing

mmatvein:
Added support for deserializing for example into a Vec<T> in a proto_bsn asset. I used the List term instead of Vec as it seemed to be used in bevy_reflect.

```
-- foo.rs
#[derive(Component, ...)]
#[reflect(...)]
pub struct Foo {
    pub values: Vec<f32>
}

-- foo.proto_bsn:
(
    Foo {
        values: [ 0.5, 1.2, 1.9, 2.6 ]
    }
)
```

NOTE: This does not yet add support for deserializing into arrays. That could be implemented by adding support for DynamicArray loading in reflect_list.


jbuehler23:
same here as well, needs some help with clippy

villor (approved):
Very useful addition. Code looks good to me.

Thank you!

PR #206, opened 2025-06-07
Fix EntityPatch and scene retention bugs in `bevy_proto_bsn`

villor:
- Fixes a bug in `EntityPatch` where scenes without inheritance would be dynamically constructed instead of statically and vice versa
- Fixes panics during scene retention when trying to despan non-existent orphans


alice-i-cecile (approved):
Very happy to see this improving :)

PR #211, opened 2025-06-12
Enhance properties pane UI with improved layout and styling for entity selection and component display

jbuehler23:
###Current work in Progress
image

Just reorganizing the Properties Panel to display more information about the selected Entity from the Scene Tree in the Properties Panel as string representations using Reflection. Early feedback welcome!

### DIFF
--- a/bevy_editor_panes/bevy_properties_pane/src/lib.rs
+++ b/bevy_editor_panes/bevy_properties_pane/src/lib.rs
@@ -51,10 +51,18 @@ fn update_properties_pane(
 fn properties_pane(selected_entity: &SelectedEntity, world: &World) -> Template {
     match selected_entity.0 {
         Some(selected_entity) => component_list(selected_entity, world),
-        None => template! {(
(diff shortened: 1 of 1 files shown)


bevy_editor_panes/bevy_properties_pane/src/lib.rs
```diff
@@ -72,12 +80,25 @@ fn component_list(entity: Entity, world: &World) -> Template {
             template! {
                 Node {
                     flex_direction: FlexDirection::Column,
+                    margin: UiRect::all(Val::Px(4.0)),
+
+                    // BackgroundColor(Color::srgb(0.2, 0.2, 0.2)),
```
alice-i-cecile, line 85: No commented out code please :)

alice-i-cecile (changes requested):
Looks nice once we remove the dead code :)

jbuehler23:
I've added some changes so that we can start to see the actual Component values rather than the field names:
image

@lmsonic

PR #212, opened 2025-06-15
State preserving template building

Constanze3:
## Summary

This change introduces a new approach to template building where components of entities that already exist (based on their anchor) are preserved, unless they are not part of the template in which case they will be despawned.

This functionality can be accessed by calling `build_preserving` or `build_children_preserving` on `EntityCommands`.

I also added a few tests to verify this behavior.

## Why?

Currently, when building templates, the entire component set of an entity is overwritten by the template, regardless of whether the entity already exists. This behavior is problematic for UI elements that rely on persistent state, as it resets the state of child components.

## Example

Consider a scenario with expansion tiles (foldable UI elements) in the scene tree. If some tiles are folded and an entity is deleted, the scene tree will need to be rebuilt. With the current approach, rebuilding the scene results in the template overwriting the state of the expansion tiles, causing them to unfold and effectively discarding the user's prior state.

### DIFF
--- a/bevy_widgets/bevy_i-cant-believe-its-not-bsn/src/template.rs
+++ b/bevy_widgets/bevy_i-cant-believe-its-not-bsn/src/template.rs
@@ -222,7 +222,6 @@ fn build_template_base(
         .into_iter()
         .map(|(fragment, anchor, entity, new)| {
             if preserving {
-                println!("HERE");
                 fragment.build_preserving(entity, world, !new);
             } else {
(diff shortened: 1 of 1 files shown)


Constanze3:
I think the implementation is fine. I’ve read through the code several times and I’m already using it in my scene tree prototype on my fork. I’ll add a few small tests later.

Constanze3:
This is not really a needed change for the editor, so am closing this to keep to open PRs clean 😅.

PR #217, opened 2025-06-29
Add list of editor planned features

Atlas16A:


alice-i-cecile (changes requested):
I like this list, but I'm not sure about the structure / how it fits into the rest of the content. At the very least you'll need to update the book index to include this page :)

design-book/src/editor_features.md
```diff
@@ -0,0 +1,183 @@
...
+Nice to have
+
+- Randomize Transform
+  - Scale
+  - Rotation
+  - Location
+- Snapping to specialized empties to assist kitbashing
+- Fast blockout tool (Blender add mesh interactively tool)
+- Mass placement tools
+- Object paining tools (a type of mass placement)
```
Jondolf, line 59:
```suggestion
- Object painting tools (a type of mass placement)
```

PR #218, opened 2025-07-08
Fixed links to the design book

YohDeadfall:
Now they're real links, not text in square brackets.


YohDeadfall:
I checked the merge queue and the pull request isn't in it. Perhaps, it's due to the unrelated CI failure.

PR #226, opened 2025-07-17
Fix 3d viewport picking passthrough and camera input

tim-blackbird:
Picking for objects through the viewport will now behave properly, including the camera controller.
It only took 8 months! <sub> well, just one day..<sub> today.

## Solution
- Assign each 3d viewport panel/camera their own custom pointer entity
- Copy every `PointerInput` event action from the mouse to these custom pointers with their location adjusted
- Rework camera input to respond to input from custom pointers

### DIFF
--- a/crates/bevy_editor_cam/src/input.rs
+++ b/crates/bevy_editor_cam/src/input.rs
@@ -79,63 +79,62 @@ pub fn default_camera_inputs(
     let pan_start = MouseButton::Left;
     let zoom_stop = 0.0;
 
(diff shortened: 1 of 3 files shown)


crates/bevy_editor_cam/src/input.rs
tim-blackbird, line 1: The changes here look big, but it's barely changed. The code was shifted around and indentation was dropped by 2.

alice-i-cecile (approved):

PR #230, opened 2025-07-18
Improve selection

tim-blackbird:
- Moved selection code to its own module
- Added helper methods to `SelectedEntity`, and some observer system handlers
- Made the plane selectable by clicking it in the scene
- Hacked together the excellently named `Pointer<DragCancelClick>` event. It only responds to clicks without pointer movement to prevent double interactions with draggable entities, or camera controllers.

### DIFF
--- a/crates/bevy_editor_core/src/selection.rs
+++ b/crates/bevy_editor_core/src/selection.rs
@@ -1,0 +1,86 @@
+//! Editor selection module.
+
+use bevy::{ecs::entity::Entities, prelude::*};
+
+/// Editor selection plugin.
+#[derive(Default)]
+pub struct SelectionPlugin;
+
+impl Plugin for SelectionPlugin {
+    fn build(&self, app: &mut App) {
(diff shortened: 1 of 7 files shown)


crates/bevy_editor_core/src/selection.rs
```diff
@@ -0,0 +1,86 @@
...
+        self.0 = Some(entity);
+    }
+
+    /// Empty the selection.
+    pub fn reset(&mut self) {
+        self.0 = None;
+    }
+}
+
+/// System to removes entity [`SelectedEntity`] when the entity is despawned.
```
JMS55, line 45:
```suggestion
/// System to reset [`SelectedEntity`] when the entity is despawned.
```

tim-blackbird: Oops, must have tried rewriting that at some point. Thanks!

jbuehler23 (approved):

PR #234, opened 2025-07-20
Humble Beginnings for the `ActionRegistry`

tim-blackbird:
This adds a registry of 'actions', also known as [commands](https://code.visualstudio.com/api/ux-guidelines/command-palette) in VS Code, or [operators](https://docs.blender.org/manual/en/4.5/interface/operators.html) in Blender.
There's also rudimentary support for keyboard shortcuts.


alice-i-cecile (approved):
Excellent; this is the right shape to get started :)

PR #240, opened 2025-08-01
Add support for selecting multiple entities

tim-blackbird:
- Change the selection from `pub Option<Entity>` to `Vec<Entity>`. The methods ensure each entity in the selection is unique.
- Introduce the idea of a 'primary selection' for tools that can only operate on one entity at a time, like the inspector (currently). This is the entity that was last selected that has not since been deselected.

Of course, the interactions here need design, but for now I've set it up to kinda~ match Godot/Blender.
Clicking sets the selection to an entity, clicking in the viewport while holding `shift`, or clicking in the scene tree while holding `ctrl` will toggle selection for an entity. Yes, they use different modifier keys.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,13 +27,13 @@ unsafe_op_in_unsafe_fn = "warn"
 unused_qualifications = "warn"
 
 [workspace.dependencies]
-bevy = { git = "https://github.com/cart/bevy.git", rev = "ab733198b8b59ec90194797f4a1d46a12043ebfc", features = [
+bevy = { git = "https://github.com/cart/bevy.git", rev = "40e6f12bbad44dfe2566fda2ead83469d53717d0", features = [
     "wayland",
     "experimental_bevy_feathers",
 ] }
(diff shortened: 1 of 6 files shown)


crates/bevy_editor_core/src/selection.rs
```diff
@@ -2,90 +2,136 @@
...
+        }
     }
 }
 
-/// The currently selected entity in the scene.
+/// The currently selected entities in the scene.
 #[derive(Resource, Default, Reflect)]
 #[reflect(Resource, Default)]
-pub struct SelectedEntity(pub Option<Entity>);
+pub struct EditorSelection(Vec<Entity>);
```
alice-i-cecile, line 45: You should use https://dev-docs.bevy.org/bevy/ecs/entity/type.UniqueEntityVec.html instead, which will move these invariants to the type level and allow for safe parallel access in the future if needed.

alice-i-cecile (changes requested):
Generally love this, but we have fancy unique entity vectors that are a perfect fit here.

JMS55:
HashSet seems more appropriate than a Vec, no?

alice-i-cecile (approved):

PR #248, opened 2025-09-05
Macos native menubar

aminwhat:
Hi there

I was working with the simple editor example and saw the menu bar on the top
kinda didn't like it on mac ;) 
so i've added the macos native menu bar, on window still uses the previous menu bar(it least it should) but on mac it uses the macos native menu bar at the top

macos-native-menubar

i didn't know if it is wanted it or not, so i've created the pull request

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,3 +84,9 @@ bevy_undo = { path = "crates/bevy_undo" }
 bevy_infinite_grid = { path = "crates/bevy_infinite_grid" }
 bevy_editor_cam = { path = "crates/bevy_editor_cam" }
 bevy_clipboard = { path = "crates/bevy_clipboard" }
+
+# MacOS Native Dependencies
+objc-rs = "0.2.8"
+objc2 = "0.6.2"
+objc2-app-kit = "0.3.1"
+objc2-foundation = "0.3.1"
--- a/bevy_widgets/bevy_menu_bar/Cargo.toml
+++ b/bevy_widgets/bevy_menu_bar/Cargo.toml
@@ -6,6 +6,11 @@ edition = "2024"
 [dependencies]
 bevy.workspace = true
 bevy_editor_styles.workspace = true
+cocoa = "0.26.1"
+objc-rs = "0.2.8"
+objc2 = "0.6.2"
+objc2-app-kit = "0.3.1"
+objc2-foundation = "0.3.1"
 
 [lints]
 workspace = true
--- a/bevy_widgets/bevy_menu_bar/Cargo.toml
+++ b/bevy_widgets/bevy_menu_bar/Cargo.toml
@@ -6,11 +6,12 @@ edition = "2024"
 [dependencies]
 bevy.workspace = true
 bevy_editor_styles.workspace = true
-cocoa = "0.26.1"
-objc-rs = "0.2.8"
-objc2 = "0.6.2"
-objc2-app-kit = "0.3.1"
-objc2-foundation = "0.3.1"
+
+objc-rs.workspace = true
+objc2.workspace = true
+objc2-app-kit.workspace = true
+objc2-foundation.workspace = true
+
 
 [lints]
 workspace = true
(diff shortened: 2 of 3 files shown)


alice-i-cecile:
For history, we've discussed this before: @mockersf likes this, I don't :)

mockersf:
> For history, we've discussed this before: @mockersf likes this, I don't :)

My view on this is that people that use macOS want consistency with their OS, and people that don't use macOS want consistency across different OSes.

This is kind of a mentality split. macOS has good support for menu bars and people used to it hate apps that reimplement it. Other systems don't, so it doesn't seem important to their users. There's also the fact that people on macOS tend to stay on macOS, while people on Windows/Linux tend to switch between the two.

Cargo.toml
```diff
@@ -84,3 +84,9 @@ bevy_undo = { path = "crates/bevy_undo" }
 bevy_infinite_grid = { path = "crates/bevy_infinite_grid" }
 bevy_editor_cam = { path = "crates/bevy_editor_cam" }
 bevy_clipboard = { path = "crates/bevy_clipboard" }
+
+# MacOS Native Dependencies
+objc-rs = "0.2.8"
```
james7132, line 89: Can you put these behind a cfg block?

bevy_widgets/bevy_menu_bar/Cargo.toml
```diff
@@ -7,5 +7,11 @@ edition = "2024"
 bevy.workspace = true
 bevy_editor_styles.workspace = true
 
+objc-rs.workspace = true
```
james7132, line 11: Same here.

james7132 (commented):
I'm in favor of it so long as it doesn't impose unreasonable restrictions while keeping menu bar behavior consistent between supported editor platforms. The current solution to split the implementation based on compilation target at the Plugin level doesn't seem tenable if we ever want to expand this.

AFAIK, Unity and Unreal's editors (not sure about Godot) support a user-scripted extension interface to their menu bars that is consistent regardless of what platform the editor runs on, and I think we should support something similar.

ZorroSvardendahl:
I'm very for having a macOS native menu bar. I like when my software feels integrated into the OS I'm using! If you use macOS you expect to have a native menu bar, I think it's more confusing for users that expect that to find it integrated into the window.

Cargo.toml
aminwhat: I'm not sure how to do it in the main Cargo.toml(in this case because of the workspace)

but i could create a crate only for macos deps and use it wherever is necessary

aminwhat: Or move the them in the bevy_menu_bar and put them there behind a cfg

it seems that workspace dependencies in the main Cargo.toml does not support this:
```
# MacOS Native Dependencies (only included when building on macOS)
[workspace.target.'cfg(target_os = "macos")'.dependencies] # or [target.'cfg(target_os = "macos")'.dependencies]
objc-rs = "0.2.8"
objc2 = "0.6.2"
objc2-app-kit = "0.3.1"
objc2-foundation = "0.3.1"
```

rust analyzer says:
```
Caused by: failed to load manifest for dependency `bevy_menu_bar` Caused by: failed to parse manifest at `/Users/aminwhat/Desktop/Projects/bevy_editor_prototypes/bevy_widgets/bevy_menu_bar/Cargo.toml` Caused by: error inheriting `objc-rs` from workspace root manifest's `workspace.dependencies.objc-rs` Caused by: `dependency.objc-rs` was not found in `workspace.dependencies
```

bevy_widgets/bevy_menu_bar/Cargo.toml
aminwhat: here it works:
```
[package]
name = "bevy_menu_bar"
version = "0.1.0"
edition = "2024"

[dependencies]
bevy.workspace = true
bevy_editor_styles.workspace = true

[target.'cfg(target_os = "macos")'.dependencies]
objc-rs.workspace = true
objc2.workspace = true
objc2-app-kit.workspace = true
objc2-foundation.workspace = true


[lints]
workspace = true
```

but i'm not sure if it fixes that mac os native deps does not pull by the rust analyzer on windows or linux

PR #257, opened 2025-10-30
Changed content directory sorting to run whenever the directory is changed, rather than when it is inserted

Freyja-moth:


Freyja-moth:
Github seems to have ran out of space to do the tests, which means the ci failing isn't my fault for once

PR #263, opened 2026-01-14
Fix ci out of space

Sieluna:
This is just a temporary patch to push the merging of multiple PRs. It is not the best solution.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,23 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
 
       - name: Install dependencies
-        run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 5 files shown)


alice-i-cecile:
Let's see if this passes CI :)

.github/workflows/ci.yaml
```diff
@@ -30,7 +30,33 @@ jobs:
...
+          sudo apt-get remove -y azure-cli google-chrome-stable firefox powershell mono-devel libgl1-mesa-dri --fix-missing || true
+          sudo apt-get remove -y google-cloud-sdk --fix-missing || true
+          sudo apt-get remove -y google-cloud-cli --fix-missing || true
+
+          # Clean up unused packages
+          sudo apt-get autoremove -y
+          sudo apt-get clean
 
       - name: Populate target directory from cache
         uses: Leafwing-Studios/cargo-cache@v2
```
TimJentzsch, line 62: You can also try replacing this with `Swatinem/rust-cache@v2` and check if it uses less space. We don't actively maintain `cargo-cache` anymore

Sieluna: I don't plan to replace workflow in this PR for now, as it could result in longer CI time. The current fix is a minor change with the goal of unblocking the stuck PR. We can open another PR for this.

PR #264, opened 2026-01-25
Update Design book architecture page

Atlas16A:
Update the architecture to reflect the current general consensus for the MVP editor. Several of the open questions have been answered at least in regards to the MVP.
First of many updates to come to the book.


design-book/src/architecture.md
```diff
@@ -5,45 +5,104 @@ But at the same time, a new user's initial experience should be polished, reason
...
+## Open questions
 
 These questions are pressing, and need serious design work.
 
-- how do we distribute the Bevy editor?
-  - do users need to have a local working copy of the Rust compiler to use the editor effectively?
-- how does the Bevy editor communicate with the Bevy game to enable effective scene editing with user-defined types?
-- how should undo-redo be handled?
+- How should undo-redo be handled?
+- How should the `editor_core` api be designed?
```
alice-i-cecile: This question is unclear.

alice-i-cecile (changes requested):
I don't like these changes: a lot of useful content has been lost, and there's a number of contentious changes. I don't have the time to review / revise this further right now, sorry Rin.

Atlas16A:
> I don't like these changes: a lot of useful content has been lost, and there's a number of contentious changes. I don't have the time to review / revise this further right now, sorry Rin.

Im not sure what content has been lost? Some stuff was just reworded and moved elsewhere in the page (line 24 into 20) and some of it is questions that the added portions answers, (line 32 answered by 76). I only see a few lines that went from explicit to implicit within the doc (like lines 20-23 got condensed down into line 85). Overall the arch is the same as what was discussed in the discord, just better worded with more clarity to reasoning. 
Do you want me to bring back the stuff that become implicit? Or is it the structure of the changes? I felt keeping the Question format didn't make much sense since it was moving towards a general consensus?

alice-i-cecile:
I would prefer to keep the structure unaltered and the changes minimal. It's quite challenging to review the diff currently.

PR #267, opened 2026-04-10
fix(gizmos): scale child positions on viewport resize + add messages

### LINKED ISSUE #253: Inconsistent transform gizmo size while scaling viewport
## Bevy version

ef28bd9

## What you did

Open the editor and resize the 3d viewport

https://github.com/user-attachments/assets/4281e192-d14e-42b4-8681-fab9b585505d

## What went wrong

The transform gizmo scales inversely with the size of the viewport

Extermind:
fix(gizmos): scale child positions on viewport resize + add messages
crate

When scaling TransformGizmo for viewport size changes, arrow tips and
plane
handles were not moving proportionally because local translations
remained
fixed while only GlobalTransform was overridden.

Solution manually scales each child's translation vector by
viewport_scale
when computing final GlobalTransform via gizmo_pos.mul_transform().

Added `bevy_transform_gizmos` crate with `ViewportSize` type and
`ViewportResized`
message for standardized event-driven viewport management across editor
systems.

Fixes #253

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,3 @@ bevy_undo = { path = "crates/bevy_undo" }
 bevy_infinite_grid = { path = "crates/bevy_infinite_grid" }
 bevy_editor_cam = { path = "crates/bevy_editor_cam" }
 bevy_clipboard = { path = "crates/bevy_clipboard" }
-
-# editor events/messages
-bevy_events = { path = "crates/bevy_events" }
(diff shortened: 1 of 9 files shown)


alice-i-cecile:
Was this intended to be marked as a draft? :)

Extermind:
Yeah, it's because I'm still new to GitHub. I primarily use a different platform for work-related code. Marked as ready for review.

alice-i-cecile (changes requested):
We definitely shouldn't have a dedicated events crate. We split code by functionality in here :)

The fix itself looks reasonable though.

